PR #285, opened 2023-06-09
allow indexing into strings

wolfv:
In the rattler-build project we came across the issue that we couldn't index into strings. Not sure if this is intentional or not, but this PR attempts to fix it (if it wasn't intentional).

https://github.com/prefix-dev/rattler-build/issues/149

I tested it and it works, but before working further on this I wanted to check if this is desired in minijinja or not.

Also `{{ name | first }}` seems to work fine for strings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to MiniJinja are documented here.
 
+## 1.0.0-alpha.2
+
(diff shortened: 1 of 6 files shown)


wolfv:
Ah, also the range syntax does work fine (so `{{ name[0:1] }}` works. Only indexing directly seems to not work.

wolfv:
thanks!

PR #286, opened 2023-06-11
Added exposed error info

mitsuhiko:
This replaces #284 but with pretty much the same intention.


nokome:
Thanks for this, works perfectly for my use case.

PR #443, opened 2024-03-22
Tolerate underscores in number literals

inducer:
In a way a follow-on to #428, needed for compatibility with Jinja2 in my use case.


minijinja/src/compiler/lexer.rs
inducer, line 1: I'm very much a Rust newbie. I'm glad I got this to go without too much trouble, but I make no promises as to whether it's idiomatic. Happy to fix things/have you push fixes.

PR #603, opened 2024-10-23
Switch to u64 types for value in cabi

mitsuhiko:
Fixes #601


PopFlamingo:
@mitsuhiko Awesome thanks for the quick fix!

PR #628, opened 2024-11-01
(Don't merge) Prepare to publish as axominijinja

fasterthanlime:
This is here just to show what we'll publish as `minijinja` to crates.io


fasterthanlime:
(Opened against the wrong fork 🙃  please ignore, sorry Armin!)

PR #649, opened 2024-11-24
added wordcount filter

jqnatividad:


yacir:
Nice!
However, I noticed that the implementation doesn’t fully match the Jinja2 result.

Using split_whitespace might fail, as Jinja2 relies on the \w+ regex. Here’s an example of a mismatched result:

```python
{{'hello-world'|wordcount}} # -> 2
{{'hello@world'|wordcount}} # -> 2
```

```rust
"{}", "hello@world'".split_whitespace().count() // -> 1
"{}", "hello@world'".split_whitespace().count() // -> 1
```

jqnatividad:
That @yacir for catching that as I just went for the naive implementation and didn't check Jinja2.

PR #681, opened 2025-01-24
feat: `MediumStrict` undefined behavior that disallows emitting undefined

wolfv:
A prototype implementation for #680 

This follows strict undefined rules for iterating and attribute access, but allows testing against undefined. It disallows emitting undefined values.

### DIFF
--- a/minijinja/src/environment.rs
+++ b/minijinja/src/environment.rs
@@ -796,7 +796,12 @@ impl<'source> Environment<'source> {
         state: &State,
         out: &mut Output,
     ) -> Result<(), Error> {
-        if value.is_undefined() && matches!(self.undefined_behavior, UndefinedBehavior::Strict) {
+        if value.is_undefined()
+            && matches!(
+                self.undefined_behavior,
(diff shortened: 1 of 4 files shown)


wolfv:
I think I might have either mis-remembered or the behavior has changed, but the strict behavior does seem to align more closely with what we would like then I had in mind.

However, this one currently fails:

```txt
{{ 'foobar' if undefvar else 'baz' }}
```

While this one passes:

```txt
{{ 'foobar' if not undefvar }}
```

wolfv:
Ok, I guess my major problem (and that I didn't solve) is that 

```
{{ "foo" if 1 > 2 }}
```

raises because the non-existent else part implicitly returns `undefined` and then it throws the error at printing time. I am not sure where I could insert a workaround for this, and how that would look like. Maybe you have some tips!

wolfv:
I think what I actually want is an `if` that doesn't return `undefined` when else evaluates, but I am unsure how to achieve that in a clean way. I think I need to think a bit longer about this.

PR #776, opened 2025-04-13
cabi: return an empty string when no debug info

maxbrunet:
Follow up #775. When debug is disabled, a `\n` newline character is still returned due to the use of `writeln!()`. 

Not sure if it is the best approach to check is debug info are set or not.

### DIFF
--- a/minijinja-cabi/src/error.rs
+++ b/minijinja-cabi/src/error.rs
@@ -60,14 +60,19 @@ pub unsafe extern "C" fn mj_err_get_debug_info() -> *mut c_char {
                 .and_then(|x| {
                     let mut info = String::new();
                     if x.name().is_some() {
-                        writeln!(info, "{}", x.display_debug_info()).unwrap();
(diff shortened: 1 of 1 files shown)


minijinja-cabi/src/error.rs
```diff
@@ -60,14 +60,19 @@ pub unsafe extern "C" fn mj_err_get_debug_info() -> *mut c_char {
...
                     if x.name().is_some() {
-                        writeln!(info, "{}", x.display_debug_info()).unwrap();
+                        let err_info = x.display_debug_info().to_string();
+                        if !err_info.is_empty() {
+                            writeln!(info, "{err_info}").unwrap();
+                        }
                     }
                     let mut source_opt = x.source();
                     while let Some(source) = source_opt {
-                        writeln!(info, "\ncaused by: {source}").unwrap();
```
maxbrunet, line 67: Not sure if that line is relevant when there's no debug info, adding it back in doubt

PR #827, opened 2025-08-19
Add multiple-key support in the sort filter

atsmtat:
Currently, the `sort` filter allows sorting objects using a custom key via the "attribute" keyword argument, which is limited to fetching a single field from the object and using that as the key. Sometimes it is useful to sort objects using multiple keys, or a composite key made of multiple fields of the object. Add a new keyword argument "keys" to support such keys.

Now, you can not only sort cities by their name ...

`{{ cities|sort(attribute="name") }}`

... but also sort cities with the same name by their country:

`{{ cities|sort(keys=["name", "country"]) }}`

### DIFF
--- a/minijinja/src/filters.rs
+++ b/minijinja/src/filters.rs
@@ -740,7 +740,7 @@ mod builtins {
     ///
     /// * `case_sensitive`: set to `true` to make the sorting of strings case sensitive.
     /// * `attribute`: can be set to an attribute or dotted path to sort by that attribute.
-    /// * `keys`: a list of attributes or dotted paths forming a composite key to sort with.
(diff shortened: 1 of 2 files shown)


atsmtat:
Thanks for taking a look!

I missed that Jinja2's sort supports multi-key via comma-separated list to the "attribute" argument: https://jinja.palletsprojects.com/en/stable/templates/#jinja-filters.sort

So I can change this filter to do the same if that sounds better to you.

Also, I was considering adding a similar "attribute" keyword arg support to `dictsort` filter, but since Jinja2 doesn't have it, I assume you'd prefer not doing that? Or would you be open to it?

PR #868, opened 2026-02-25
Add 'c' format type support to format filters

tomzie-arista:
Add 'c' character formatting type to both styles of the format filter.
Follow closely what Python supports.

_AI was used to generate the tests._

### DIFF
--- a/minijinja/src/format_utils.rs
+++ b/minijinja/src/format_utils.rs
@@ -410,13 +410,37 @@ impl FormatSpec {
                         format!("{} arg not in range(0x110000)", self.ty.description()),
                     ));
                 }
-                if self.format_style == FormatStyle::StrFormat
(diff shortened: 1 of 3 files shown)


minijinja/src/format_utils.rs
```diff
@@ -386,6 +403,29 @@ impl FormatSpec {
...
+                        ErrorKind::InvalidOperation,
+                        "%c arg not in range(0x110000)",
+                    ));
+                }
+                if self.format_style == FormatStyle::StrFormat
+                    && (self.print_sign || self.space_before_positive_num)
+                {
+                    return Err(Error::new(
+                        ErrorKind::InvalidOperation,
+                        "sign flags are not allowed with %c format type",
```
atsmtat: "%c format type" text in the message sounds wrong for str.format. Use `self.ty.description()` to print the right description.

Issue #876, opened 2026-03-02
Error on undefined variables that are the input of a filter


Hofer-Julian:
This is a simple minijinja setup that runs to completion without panicking.
Notice the `UndefinedBehavior::Strict` at the beginning of the function.

```rust
fn main() {
    let mut env = minijinja::Environment::new();
    env.set_undefined_behavior(minijinja::UndefinedBehavior::Strict);

    let ctx = minijinja::context! {};

    // This behaves as I expect it
    let result = env
        .render_str("{{ x | default('FALLBACK') }}", &ctx)
        .unwrap();
    assert_eq!(result, "FALLBACK");

    // I would expect this to panic
    let result = env.render_str("{{ x | upper }}", &ctx).unwrap();
    assert_eq!(result, "");
}
```

While the first example works as expected, I would that the second example panics. At the moment, the undefined variable is just accepted and converted to an empty string.

Would you be interested in changing the default to erroring in that case as well.
I currently implement the behaviour like [this](https://github.com/prefix-dev/rattler-build/pull/2239), but l'd love to upstream the behaviour if possible.

PR #877, opened 2026-03-02
fix: complain when filter is used on undefined in strict and semi-strict mode

wolfv:
This fixes https://github.com/mitsuhiko/minijinja/issues/876

AI Disclaimer: Code entirely written by Claude.


wolfv:
Thanks! A release would be amazing :)

PR #900, opened 2026-05-16
fix(lexer): support mid-chain dotted-integer attribute access

KrishnanPrash:
#### Overview

Jinja2 lets you index a sequence with a dotted integer (`obj.0` is sugar for `obj[0]`), and the result composes with attribute access (`obj.0.attr`). minijinja's tokenizer was eating the second dot as the start of a fractional part, so the parser saw a `Float` where it expected an identifier or integer.

Before: `{{ msgs.0.role }}` -> `syntax error: unexpected float, expected identifier or integer`
After:  renders the attribute value.

#### Behavior

After an integer, the lexer now classifies a `.` by the byte(s) that follow it:

| After the dot | Treat as | Example |
| --- | --- | --- |
| digit | fraction | `1.5` |
| `[eE][+-]?\d` | exponent | `1.e5` |
| identifier-start | attribute access (break) | `obj.0.attr` |
| else (ws / op / EOF) | fraction (unchanged) | `42.` |

The predicate reuses `lex_identifier` so it tracks the identifier grammar under both ASCII-only and `unicode` feature configs. The parser's `parse_subscript` already accepts an integer after `.`.

#### Tests

- `cargo test -p minijinja --all-features`: 19 binaries, 0 failures.
- Float-literal regressions pinned for `1.0`, `1.5e2`, `1.e5`, `1.E-3`, `42.`, `1.+2`.
- Attribute cases: single-digit, multi-digit, `1_000`, `_meta`, Greek `α` under `unicode`.
- `test_expression_bug` was split since its example (`42.blahadsf()`) is now well-formed.

#### Reviewer

Start with the `(b'.', State::Integer)` arm of `eat_number` in `minijinja/src/compiler/lexer.rs`.

#### AI Disclosure

Developed with AI assistance (Claude Code), including this description. I reviewed and verified.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3358,9 +3358,9 @@ dependencies = [
 
 [[package]]
 name = "toml"
-version = "0.7.8"
+version = "0.8.23"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "dd79e69d3b627db300ff956027cc6c3798cef26d22526befdfcd12feeb6d2257"
+checksum = "dc1beb996b9d83529a9e75c17a1686767d148d70663143c7854d8b4a09ced362"
 dependencies = [
  "serde",
  "serde_spanned",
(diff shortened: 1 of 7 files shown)


KrishnanPrash:
@mitsuhiko Would appreciate a review for this PR whenever you get the chance. Thank you in advance.

rmccorm4:
Hi @mitsuhiko, is this something you could help us get merged and put into a new crate release that we can update upstream?

PR #902, opened 2026-05-26
feat(environment): add auto escape callback for templates

qlux:
Expose the autoescape parameter to the js binding.

---
Co-Authored-By: Claude Opus <noreply@anthropic.com>

### DIFF
--- a/minijinja-js/src/lib.rs
+++ b/minijinja-js/src/lib.rs
@@ -236,6 +236,33 @@ impl Environment {
                 }
             });
     }
+
+    /// Sets a callback to determine the auto escaping behavior for a template.
+    ///
+    /// The callback receives the template name and should return one of the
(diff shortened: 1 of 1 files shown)


mitsuhiko (changes requested):
Thanks! The core `html`/`json`/`none` behavior works locally, but I don't think this is ready to merge yet:

- The custom-format branch leaks a new allocation on every callback invocation via `Box::leak`. The JS binding does not expose a custom formatter, so this advertised mode cannot actually render and currently just errors. Please restrict the return values to the supported modes (or otherwise avoid the per-call leak).
- Please add JS tests covering the callback/name and escaping behavior.
- Please update `minijinja-js/README.md`, which still lists custom auto-escape callbacks as unsupported.

I verified that the branch compiles and that the existing JS test suite passes after merging current `main`.

PR #906, opened 2026-06-09
Improve performance of iteration over map items

atsmtat:
The Enumerator enum underlying the object iteration support in the engine currently
allows only one kind of iterator yielding a single Value as the item. So the map-like
objects currently store an iterator over the keys in the Enumerator. This means, in
contexts where both key and value are required, e.g. items and dictsort filters, the
renderer has to call `get_value` to fetch the value corresponding to the key while
iterating the map. This extra lookup is wasteful, and can be avoided if the
Enumerator can store an iterator of the key value pairs of a map.

Expand the enum with `KeyValueIter` and `RevKeyValueIter` alternatives storing an
iterator over key-value pairs. Update `Object::try_iter` to return an iterator over
the keys for the new enumerators in case the ObjectRepr is Map, to avoid breaking
their iteration behavior in templates. Update `Object::try_iter_pairs` to return the
underlying iterators directly for the new enumerators.

Add two new helper methods in the ObjectExt trait to allow creating new enumerators
projecting into the given object, similar to `mapped_enumerator` method. Update docs
on the existing methods. Fix a wrong example on the `mapped_rev_enumerator` method.

Update Object impl of HashMap, IndexMap, and BTreeMap to use the new enumerators.

Implement Object trait for arrays, and use array instead of vector to store key value pairs
to reduce the number of memory allocations.

Add a benchmark for iterating over a map. Results on MacBook Pro M1:
```
loop_map_items          time:   [2.3926 ms 2.3958 ms 2.3993 ms]
                        change: [-48.584% -48.480% -48.377%] (p = 0.00 < 0.05)
                        Performance has improved.
```

(No AI assistance was used to prepare these changes, everything is hand-crafted).

### DIFF
--- a/minijinja/src/value/object.rs
+++ b/minijinja/src/value/object.rs
@@ -346,7 +346,7 @@ macro_rules! impl_object_helpers {
                     if let ObjectRepr::Map = self.repr() {
                         Some(Box::new(iter.map(|(key, _)| key)))
                     } else {
-                        Some(Box::new(iter.map(|(key, val)| vec![key, val].into())))
+                        Some(Box::new(iter.map(|(key, val)| [key, val].into())))
                     }
                 }
                 Enumerator::RevIter(iter) => Some(Box::new(iter)),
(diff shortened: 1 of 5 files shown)


minijinja/src/value/object.rs
```diff
@@ -1193,7 +1193,7 @@ where
     }
 }
 
-impl<T, const N : usize> From<[T; N]> for Value
+impl<T, const N: usize> From<[T; N]> for Value
```
jonathanc-arista, line 1196: Nit: move this formatting fix into the previous commit?

jonathanc-arista (changes requested):

PR #907, opened 2026-06-10
fix: avoid stack overflow on repeated sequence concatenation

rmccorm4:
### Overview

Repeated `seq = seq + [x]` (e.g. a chat-template accumulator over many
turns) built an unbounded chain of lazy iterables. Each `+` wrapped the
previous result in another `make_iterable` closure, so later enumeration
of the value (for example `{{ seq | length }}` -> `Value::len()`)
recursed one native stack frame per concatenation and overflowed the
stack for long chains, aborting the process.

### Solution

Concatenation of two sized sequences now materializes eagerly into a
flat `Seq`, matching CPython/Jinja2 `list + list` semantics, so the
result stays shallow and the chain cannot form. Unsized iterables keep
the existing lazy-chaining behavior to preserve their laziness.

#### minijinja vs jinja2 on the failing case

A template that accumulates with `+` in a loop, then takes `|length`. With
enough turns (~1500+ on a 2 MB stack), minijinja core-dumps; jinja2 succeeds.

**Shared template** (`seq = seq + [x]` accumulator → `|length`):
```jinja
{%- set ns = namespace(messages=[]) -%}
{%- for m in messages -%}{%- set ns.messages = ns.messages + [m] -%}{%- endfor -%}
{{ ns.messages | length }}
```

**minijinja — FAILS (stack overflow / SIGABRT)**
```rust
use minijinja::{context, Environment, Value};

env.add_template("chat", SRC).unwrap();
let msgs: Vec<Value> = (0..3000).map(Value::from).collect();
// renders on a ~2 MB worker stack:
std::thread::Builder::new().stack_size(2 << 20).spawn(move || {
    env.get_template("chat").unwrap()
        .render(context! { messages => Value::from(msgs) })   // <-- aborts here
}).unwrap().join().unwrap();
// thread '<unknown>' has overflowed its stack / fatal runtime error: stack overflow
```

**jinja2 — PASSES**
```python
from jinja2 import Environment

msgs = [{"role": "user"} for _ in range(3000)]
print(Environment().from_string(SRC).render(messages=msgs))   # -> "3000"
# works even with sys.setrecursionlimit(1000)
```

### AI Disclosure

Developed with AI assistance (Claude Code), including this description. I reviewed and verified.

### DIFF
--- a/minijinja/src/value/merge_object.rs
+++ b/minijinja/src/value/merge_object.rs
@@ -1,4 +1,5 @@
 use std::collections::BTreeSet;
+use std::fmt;
 use std::sync::Arc;
 
 use crate::value::ops::LenIterWrap;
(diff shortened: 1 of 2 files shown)


rmccorm4:
Thanks for the continued support on this project @mitsuhiko, we ran into a stack overflow crash from this issue for chat templates being given a conversation with many turns, which should be resolved by this fix. After this is merged, it would be great to get a new release we can upgrade to in our dependencies as well. Let me know if there's anything I can do to help.

CC @KrishnanPrash

rmccorm4:
Hi @mitsuhiko, could you help take a look here?

mitsuhiko:
I want to try an in-between solution first before eagerly evaluating iterators.

PR #913, opened 2026-07-27
fix: render None and booleans like Jinja2

### LINKED ISSUE #911: Null values render as none instead of Jinja2's None
MiniJinja renders a null value as `none`, while Jinja2 renders it as `None`.

## Reproduction

Template:

```jinja
{{ value }}
```

Context:

```json
{"value": null}
```

MiniJinja 2.21.0 outputs:

```text
none
```

Jinja2 3.1.6 outputs:

```text
None
```

This affects the published GLM-5.2 chat template. Its `visible_text` macro
directly interpolates content that is not a string or sequence, so null message
content produces different prompt bytes and token IDs.

[Source template on Hugging Face](https://huggingface.co/zai-org/GLM-5.2/blob/main/chat_template.jinja)

Expected behavior: match Jinja2 and render `None`.

michaelfeil:
Fixes #911.

Render `none`, `true`, and `false` as `None`, `True`, and `False` to match Jinja2. This applies consistently to direct output, nested values, HTML-escaped output, string coercion, and printf-style formatting in both the Rust and Go implementations. Updates the Rust/Go snapshots and the C, Python, and JavaScript binding expectations, with focused regression coverage.

Tested with `make test`, `cargo test -p minijinja --all-features`, rustfmt, the workspace clippy target, the C ABI tests, and `go test ./...`. Python and JavaScript binding suites are also covered by CI.

AI disclosure: The original fix was prepared with Codex assistance following a user-directed fix and reviewed by the contributor before marking the PR ready. The maintainer follow-up expanded the fix to booleans, all rendering paths, Go, and the bindings, and was reviewed and tested before merge.


michaelfeil:
pr is ready

michaelfeil:
cc @mitsuhiko thx for review

michaelfeil:
@codex review

PR #918, opened 2026-08-07
MiniJinja 3

mitsuhiko:
This is not to be merged, at least not as a pull request.

- Optional, removable serde dependency with fallback conversions (#528)
- First-class tuples across Rust, Python, and Go.
- Tuple-preserving slicing, concatenation, repetition, filters, and conversions.
- Jinja2-compatible collection rendering, quoting, and tojson spacing.
- Explicit keyword-argument handling via `ValueOrKwargs`.
- Move to `&mut State` as public interface.
- MiniJinja-Go migrated to module path `/v3`.

The branch is live, and more changes will land.


mitsuhiko:
Main is now minijinja 3

PR #921, opened 2026-08-11
Allow conditional expressions in keyword arguments

jbg:
Jinja permits conditional expressions as keyword argument values, but MiniJinja currently rejects them unless parenthesized. Parse keyword values as full expressions and cover the valid syntax with a rendering test.

Tested with `cargo test -p minijinja --all-features`.

AI disclosure: OpenAI Codex helped investigate the parser difference, prepare the change and description, and run the tests. The contributor reviewed the result.


mitsuhiko:
Duplicate.

PR #922, opened 2026-08-11
fix(parser): support conditional expressions in keyword arguments

KrishnanPrash:
#### Overview

Jinja2 accepts a conditional expression as a keyword-argument value; minijinja rejects it while accepting the same expression in positional position. Every kwarg site is affected (function calls, filters, tests, macro calls, `namespace()`) since they share `parse_args`.

Before: `{{ namespace(name=x if x else 'd') }}` -> `syntax error: unexpected identifier, expected ','`
After: renders.

#### Behavior

The parser has two expression entry points. `parse_expr` accepts a trailing conditional (`x if c else y`). `parse_expr_noif` does not; it exists for statements where a trailing `if` already means something else, like the loop filter in `{% for x in seq if cond %}`.

Inside a call, positional and splat arguments use `parse_expr`, but the keyword-argument value used `parse_expr_noif`, so the parser stopped at `if` and expected a `,`. An argument list is wrapped in parentheses and separated by commas, so a trailing `if` there can only start a conditional. The fix switches the keyword-argument value to `parse_expr`. The statement sites keep `parse_expr_noif`. Jinja2 draws the same line (`with_condexpr` is true for call arguments, false only in `parse_for`/`parse_if`). Both the Rust and Go parsers get the fix.

#### AI Disclosure

AI (Claude Code) was used in the implementation, testing, and writing of this PR. I reviewed and verified.


KrishnanPrash:
Hey @mitsuhiko,

Could you review this PR when you get a chance? Additionally, when merging, could you put this into a new crate release so that we can update upstream? 🙂

mitsuhiko:
Resolved on minijinja-2.

PR #923, opened 2026-08-11
Make context! preserve key order with preserve_order

### LINKED ISSUE #920: context! ignores preserve_order and sorts keys
## Description

With the `preserve_order` feature enabled, values built with the `context!` macro still come out with their keys sorted. Values built with `Value::from_serialize` preserve insertion order in the same build, so the two disagree.

The cause looks like `__context::make()` in `minijinja/src/macros.rs`, which returns a `std::collections::BTreeMap` unconditionally. That map is built and then handed to `Value::from_object`, so it never goes through the indexmap-backed value map that `preserve_order` switches on.

The feature is documented in `lib.rs` as "When enable the internal value implementation uses an indexmap which preserves the original order of maps and structs", and the `context!` docs do not mention an exception, so I expected the two to agree.

This only matters when the order is observable in the output. I hit it rendering Hugging Face chat templates, where a tool definition goes through `tojson` and the result has to match Python's `json.dumps` byte for byte. Sorted keys there change the prompt a model receives.

## Reproduction steps

Cargo.toml:

```toml
[dependencies]
minijinja = { version = "2.23", features = ["preserve_order"] }
serde_json = { version = "1", features = ["preserve_order"] }
```

main.rs:

```rust
use minijinja::{context, Value};

fn main() {
    let from_macro = context! { zebra => 1, apple => 2 };
    println!("context!:       {from_macro:?}");

    let json: serde_json::Value = serde_json::from_str(r#"{"zebra":1,"apple":2}"#).unwrap();
    println!("from_serialize: {:?}", Value::from_serialize(&json));
}
```

Output:

```
context!:       {"apple": 2, "zebra": 1}
from_serialize: {"zebra": 1, "apple": 2}
```

- Version of minijinja: 2.23.0
- Version of rustc: 1.97.1 (8bab26f4f 2026-07-14)
- Operating system and version: Ubuntu 22.04.5 LTS, kernel 6.8.0

## What did you expect

I expected `context!` to keep insertion order when `preserve_order` is on, so that it agrees with `Value::from_serialize`.

If the `BTreeMap` is a deliberate choice, then a note on the feature or on the `context!` docs would have saved me the detour, since the current wording reads as covering all maps. I am happy with either outcome and can send a PR for whichever you prefer.

MsfPablo:
Fixes #920.

`__context::make()` in `minijinja/src/macros.rs` built the context in a `std::collections::BTreeMap` unconditionally, so `context!` came out with sorted keys even with `preserve_order` enabled, disagreeing with `Value::from_serialize` in the same build.

Changes:

- `__context` now uses a feature-gated `ContextMap` alias: `BTreeMap<&'static str, Value>` by default, `indexmap::IndexMap<&'static str, Value>` under `preserve_order`.
- Added `impl_str_map!(IndexMap, ...)` alongside the existing `impl_value_map!(IndexMap, ...)` in `value/object.rs`, so `IndexMap` with string keys gets the same `Object`/`From<...>` impls `BTreeMap` and `HashMap` already have. No behavior change without the feature.
- Regression test in `tests/test_macros.rs` behind `#[cfg(feature = "preserve_order")]`, plus a CHANGELOG entry.

Scope note: `context!{ a => 1, ..other }` goes through `MergeDict`, whose `enumerate` collects keys into a `BTreeSet` and therefore still sorts under `preserve_order`. That is a separate code path with its own ordering question (which source's order wins), so I left it alone rather than bundle a design decision into this fix. Happy to follow up if you want it changed.

Verification (all from this branch):

```
$ cargo fmt --all --check                # clean
$ cargo test -p minijinja                # all suites pass
$ cargo test -p minijinja --all-features # all suites pass, incl. new test
$ cargo clippy -p minijinja --all-targets --all-features -- -D warnings
```

Clippy reports one `needless_borrow` at `minijinja/tests/test_macros.rs:113` (`m.call(&state, args!(42))`). I confirmed against a `git stash` baseline that it is pre-existing on `main` and untouched by this change, so I left it as-is.

## AI disclosure

Per CONTRIBUTING.md: this contribution was made with AI assistance. The diagnosis, the code change, the test, and this PR description were all produced by an AI agent (Claude) and reviewed by me before submitting. The verification commands above were actually executed, not assumed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ All notable changes to MiniJinja are documented here.
 
 ## Unreleased
 
+* Fixed `context!` sorting its keys when the `preserve_order` feature is enabled.  #920
+
 ## 2.23.0
 
 * Fixed Unicode identifiers in templates rendered through the Python bindings.
--- a/minijinja/src/macros.rs
+++ b/minijinja/src/macros.rs
@@ -25,21 +25,26 @@ macro_rules! some {
 pub mod __context {
     use crate::value::Value;
     use crate::Environment;
-    use std::collections::BTreeMap;
     use std::rc::Rc;
 
+    #[cfg(not(feature = "preserve_order"))]
+    pub type ContextMap = std::collections::BTreeMap<&'static str, Value>;
+
+    #[cfg(feature = "preserve_order")]
+    pub type ContextMap = indexmap::IndexMap<&'static str, Value>;
+
     #[inline(always)]
-    pub fn make() -> BTreeMap<&'static str, Value> {
-        BTreeMap::default()
+    pub fn make() -> ContextMap {
+        ContextMap::default()
     }
 
     #[inline(always)]
-    pub fn add(ctx: &mut BTreeMap<&'static str, Value>, key: &'static str, value: Value) {
+    pub fn add(ctx: &mut ContextMap, key: &'static str, value: Value) {
         ctx.insert(key, value);
     }
 
     #[inline(always)]
-    pub fn build(ctx: BTreeMap<&'static str, Value>) -> Value {
+    pub fn build(ctx: ContextMap) -> Value {
         Value::from_object(ctx)
     }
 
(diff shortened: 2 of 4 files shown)


mitsuhiko:
I'll fix this differently.

PR #930, opened 2026-08-28
fix(filters): match Jinja2 word boundaries in title

jaideeppyne:
The `title` filter treats every ASCII punctuation character as the start of a new word, so `{{ "don't stop believing"|title }}` renders as `Don'T Stop Believing`.

Jinja2 splits words on `_word_beginning_split_re = re.compile(r"([-\s({\[<]+)")` — only whitespace and `-`, `(`, `{`, `[`, `<` begin a new word. This restricts the boundary set to match, which leaves the existing hyphen/bracket behavior untouched.

| template | before | after / Jinja2 3.1.6 |
|---|---|---|
| `{{ "don't stop believing"\|title }}` | `Don'T Stop Believing` | `Don't Stop Believing` |
| `{{ "foo_bar.baz"\|title }}` | `Foo_Bar.Baz` | `Foo_bar.baz` |
| `{{ "john.doe@example.com"\|title }}` | `John.Doe@Example.Com` | `John.doe@example.com` |
| `{{ "foo (bar) [baz]"\|title }}` | `Foo (Bar) [Baz]` | `Foo (Bar) [Baz]` (unchanged) |

The Go port had a third, different boundary set (`- _ : , .`), aligned here too. The existing `title-sentence` snapshot changes (`The:Word` → `The:word`), which is the intended correction.

Verified with a 400-case differential fuzz against Jinja2 3.1.6: 0 divergences across 416 total cases. `make test`, `make check`, `make lint` and `go test ./...` all pass. Happy to drop the Go hunk or split it out if you'd prefer this scoped to the Rust crate.

**AI disclosure:** this contribution was developed with AI assistance (Claude). The bug was found by differential-testing minijinja against Python Jinja2 3.1.6; the code change, the tests and this PR description were AI-assisted and human-reviewed before submission.


mitsuhiko (approved):
Reviewed against Jinja2’s `do_title` implementation and upstream title-filter cases. The boundary set and resulting behavior match (`[-\\s({\\[<]+`), and the Rust fixture tests plus Go coverage are appropriate. Local core, feature-matrix, Go, formatting, and focused Clippy checks pass.

PR #933, opened 2026-09-04
fix(go): treat empty iterators as empty sequences

andared:
In the Go port an empty iterator is not treated as an empty sequence: iterating it fails, it is truthy, and `is iterable` reports `false`. Rust MiniJinja and Jinja2 both handle these cases silently.

```jinja
{% for i in range(0, 0) %}{{ i }}{% else %}empty{% endfor %}
{% if range(0, 0) %}truthy{% else %}falsy{% endif %}
{{ range(0, 0) is iterable }}
```

| Template | Go port | Rust / Jinja2 |
|---|---|---|
| `{% for i in range(0, 0) %}` | `invalid operation: iterator is not iterable` | renders the `else` branch |
| `{% if range(0, 0) %}` | `truthy` | `falsy` |
| `{{ range(0, 0) is iterable }}` | `False` | `True` |

The same applies to a `Value` built from a nil slice: `value.FromSlice(nil)` renders `sequence is not iterable` in a `for` loop, while `value.FromSlice([]value.Value{})` iterates as expected. That distinction is invisible in a template, and a macro returning "nothing" hits it easily.

## Cause

`Value.Iter()` returns `nil` both for values that cannot be iterated and for iterable values that happen to be empty, and every caller uses `items == nil` to mean "not iterable" (the `for` loop in `state.go`, `TestIterable` in `tests/tests.go`, and about twenty filters). An empty `*Iterator` — produced by `range()` when the range is empty — therefore looks non-iterable. `Value.IsTrue()` has a case for `[]Value` but none for `*Iterator`, so an empty iterator falls through to the `default: return true`.

## Change

* `Value.Iter()` now returns an empty non-nil slice for iterable values, keeping `nil` for values that genuinely cannot be iterated. The contract is documented on the method.
* `Value.IsTrue()` treats an empty `*Iterator` as falsy, matching `[]Value`.

This also fixes `is iterable` and the affected filters, since they all key off the same `nil` check.

## Tests

Added cases to `TestRangeFunction` and `TestValueTruthiness`, plus `TestIterEmptyIsNotNil` covering the `Iter()` contract for both empty iterables and non-iterable values. `go test -race ./...` passes in `minijinja-go`; all new cases fail without the fix.

Targeting `minijinja-2` since the Go port ships as v2.24.0; the Rust implementation is unaffected.

## AI disclosure

Per CONTRIBUTING.md: this contribution was made with AI assistance. I used Claude Code (Claude Opus) to investigate the behaviour against Rust MiniJinja and Jinja2, write the patch and the tests, and draft this description; I directed the work and ran the test suite. Please review accordingly.


mitsuhiko:
Fixed directly on `minijinja-2` in 5cad94f7, with co-author credit preserved. Thank you!

Issue #903, opened 2026-06-02
Python dict methods are shadowed by same-named map keys in `Object::call_method`


BugenZhao:
## Description

Hi, and thanks for the great work first! I think there is a Python Jinja compatibility edge case around map-like objects and same-named dict methods.

In Python Jinja2, dotted lookup prefers attributes over items, so a dict key named `items` does not shadow the `dict.items` method:

```python
import jinja2

env = jinja2.Environment()
tmpl = env.from_string("{{ d.items()|list }} / {{ d['items'] }}")
print(tmpl.render(d={"items": "FIELD", "x": 1}))
```

Output:

```text
[('items', 'FIELD'), ('x', 1)] / FIELD
```

In MiniJinja, the default `Object::call_method` implementation appears to call `get_value(method)` first and then tries to call that value:

```rust
if let Some(value) = self.get_value(&Value::from(method)) {
    return value.call(state, args);
}
Err(Error::from(ErrorKind::UnknownMethod))
```

That means for a map-like object with an `items` key, `{{ d.items() }}` resolves `d["items"]` first, so we'll always enter the `value.call` path. As a result, the Python-compatible dict method fallback (registered with `minijinja_contrib::pycompat::unknown_method_callback`) cannot run, because the object never returns `UnknownMethod`.

## What did you expect

Better compatibility with Python jinja behavior:

- `d.items()` should call the dict/map method even when `d` has a key named `"items"`.
- `d["items"]` should access the key value.

Here are some different directions I can think of to address this issue:

- Enrich the current method callback to provide a list of supported methods without actually calling them, so that we can prefer Python dict methods (`items`, `keys`, `values`, `get`, etc.) over same-named map keys for method calls.

- Match the variant of the `Value` and only call it if it's "callable".

- Add a clear error variant for "not callable" and match the `Result` from `value.call(state, args)` here to determine whether to emit a `ErrorKind::UnknownMethod` instead.

mitsuhiko:
Fixed in be80b3c0c4387fcb0504aec584ea1f1734632b41 by resolving object methods and the unknown method callback before falling back to callable same-named map items.

BugenZhao:
Thanks @mitsuhiko! The fix is pretty useful and I'm bumping to 2.22 in https://github.com/vllm-project/vllm/pull/51235

PR #934, opened 2026-09-04
feat(go): add unknown method callback

andared:
Jinja2 templates routinely call Python methods on values — `s.split("-")`, `d.items()`, `s.upper()`. Rust MiniJinja does not implement those either, but it offers `Environment::set_unknown_method_callback` (#441) as the extension point, and `minijinja-contrib::pycompat` builds a full set of Python methods on top of it. The Go port has no equivalent: `MethodCallable` only covers objects the embedder writes itself, so there is no way to make `"a-b".split("-")` work for values that come out of the template context.

Without the hook the only workaround is rewriting template sources before compiling them, which is exactly the kind of thing the hook exists to avoid.

## Change

* Added `UnknownMethodFunc` and `Environment.SetUnknownMethodCallback`, matching the Rust API. Returning `value.ErrUnknownMethod` leaves the call unresolved so the regular lookup continues; any other error is reported to the caller as is.
* Extracted method resolution into `State.evalMethodCall`, which resolves `obj.name(...)` in the same order as Rust: methods implemented by the value, then the unknown method callback, then a callable stored under that name.

Resolution moved ahead of the generic callable lookup so that methods take precedence over same-named items, the way they do in Jinja2 — the ordering that #903 fixed on the Rust side. One behaviour change follows from that: a value implementing `MethodCallable` now wins over a callable attribute of the same name; previously the attribute was called first. Nothing in the existing suite depended on that.

Undefined values are left to the regular expression evaluation, so the configured undefined behaviour still applies to `{{ missing.method() }}`.

## Tests

`unknown_method_test.go` covers a callback implementing `upper`/`split`, unresolved methods, an error from the callback reaching the caller, precedence over a same-named item, and positional plus keyword arguments arriving at the callback. `go test -race ./...` passes in `minijinja-go`.

Targeting `minijinja-2` since the Go port ships as v2.24.0; the Rust implementation is unchanged.

I can follow up with a port of `minijinja-contrib::pycompat` on top of this hook if you want the Python methods available out of the box in Go.

## AI disclosure


mitsuhiko:
Implemented directly on `minijinja-2` in 7e952a39, with co-author credit preserved. Thank you!

PR #935, opened 2026-09-04
fix(go): use the euclidean remainder for integer modulo

andared:
The Go port computes the modulo of two integers with Go's `%`, which takes the sign of the dividend. Rust MiniJinja uses `math_binop!(rem, checked_rem_euclid, %)`, so on the Rust side the integer case is the euclidean remainder and only the float case uses `%`:

```jinja
{{ -7 % 3 }}
```

Rust MiniJinja and Jinja2 both render `2`, the Go port renders `-1`.

The two operators also disagree with each other inside the Go port: `FloorDiv` already floors (`{{ -7 // 2 }}` is `-4`), so `//` and `%` no longer describe the same division.

## Change

* `Value.Rem` computes the euclidean remainder for integers through a small `remEuclid` helper, matching `checked_rem_euclid`.
* Floats keep `math.Mod`, which is what Rust's f64 branch does with `%`.

## Tests

`TestRemEuclidean` covers all four sign combinations, exact division and the float case; `TestRemByZero` keeps the existing error. `go test -race ./...` and `make lint` pass in `minijinja-go`.

## AI disclosure


mitsuhiko:
Implemented directly on `minijinja-2` in 8356b5f3, with Rust and Go aligned to Jinja2 semantics and co-author credit preserved. Thank you!

PR #936, opened 2026-09-04
fix(go): keep loop.changed() state across iterations

andared:
`loop.changed()` always reports a change in the Go port:

```jinja
{% for x in [1, 1, 2] %}{{ loop.changed(x) }} {% endfor %}
```

renders `True True True`, where Jinja2 and Rust MiniJinja render `True False True`. Grouping headers, the idiom the helper exists for, therefore repeat on every row.

The callable itself is correct: it compares the arguments against the value remembered on the loop object. The problem is that the loop object is rebuilt on every iteration, so the memory never survives. Rust keeps one `Loop` per loop with a mutex-guarded `last_changed_value`; the Go port stores `changed *value.Value` on an object that is thrown away at the end of each pass.

No input in `minijinja/tests/inputs` calls `loop.changed`, which is why the shared corpus did not catch it.

## Change

* Added a small `changedMemo` that belongs to the loop as a whole and is handed to the per-iteration loop objects.
* Applied it to all three places that build a loop object: the plain sequence loop, the recursive loop body, and the pull-iterator loop. Each recursion level and each nested loop allocates its own memo, so inner loops keep their own memory.

## Tests

`loop_changed_test.go` covers repeated values, several arguments, the grouping-header idiom, nested loops keeping separate memory, and an iterator without a known length (the pull-iterator path). `go test -race ./...` and `make lint` pass in `minijinja-go`.

## AI disclosure


mitsuhiko:
Implemented directly on `minijinja-2` in 3a301d96, with shared Rust/Go regression coverage and co-author credit preserved. Thank you!

PR #937, opened 2026-09-04
fix(go): keep context floats from turning into integers

andared:
`fromReflectValue` turns any float without a fractional part into an integer, "for consistency with JSON parsing". That makes the same number render differently depending on where it came from:

```jinja
{{ 4.0 }}        {# literal:      4.0 #}
{{ x }}          {# x = float64(4): 4  #}
```

`{{ x is float }}` is then false and `{{ x is integer }}` true, and `Value.String` never gets to use its own whole-float formatting (`%.1f`), which exists precisely to match Jinja2. Rust MiniJinja keeps an `f64` an `f64` through serde, so the Go port diverges from both Rust and Jinja2 here.

The reason behind the conversion is real, though: `encoding/json` decodes every JSON number into a `float64`, so without it `{"n": 1}` would render `1.0`. Go's answer to that is `json.Decoder.UseNumber`, which the port did not support.

## Change

* `fromReflectValue` keeps `float32`/`float64` as floats.
* `FromAny` and the reflection path understand `json.Number`: a literal that parses as an integer becomes an integer, everything else a float. A decoder that cares about the distinction can now preserve it.
* `internal/testutil` decodes the JSON header of the shared test inputs with `UseNumber`, so the Rust corpus keeps seeing integers where the input file wrote integers.

Embedders that pass JSON-decoded contexts and relied on whole floats rendering as integers need the same `UseNumber` switch; that is the behaviour change here and the reason it targets a minor release.

## Tests

`value/from_any_test.go` covers floats and `json.Number` (including inside containers), `float_context_test.go` covers rendering, arithmetic and tests over context values plus both JSON literal forms. The whole shared corpus (`TestTemplates`) still matches the Rust snapshots. `go test -race ./...` and `make lint` pass in `minijinja-go`.

## AI disclosure


mitsuhiko:
Rebuilt and merged as 42c01852. Thank you!

PR #941, opened 2026-09-14
add set_syntax_callback to allow variable syntax

### LINKED ISSUE #939: Syntax should not be part of the environment but of the templates
I have to template different template files with different delimiters.  ATM I am using two Environments.  Ideally the syntax should be configurable in the file itself.

I would propose that the (a different) loader can return `Result<Option<(String, Option<Syntax>)>, Error>` instead of `Result<Option<String>, Error>`.

What do you think?

ibotty:
This was generated with the help of a clanker.  I looked at the diff myself.
I don't have the expertise to vet any performance impacts the additional `Arc<SyntaxConfig>` has.

---

This commit adds `Environment::set_syntax_callback`, guarded by the feature `custom_syntax`.

The callback is a function that takes the template's name and source and returns a `SyntaxConfig`.

Fixes #939


ibotty:
I let the clanker add a benchmark for the custom_syntax feature that shows that there is no real impact.

I am not convinced, that it tests all possible performance impacts.  Also I am not sure it belongs in the repository.

PR #954, opened 2026-09-21
fix(formatting): prevent std::fmt panic on large precision

### LINKED ISSUE #946: format filter panics on large precision — {{"%.65536f"|format(1.0)}}
## Description

$ echo '{{"%.65536f"|format(1.0)}}' | minijinja-cli -
thread 'main' panicked at .../formatting.rs:577:31:
Formatting argument out of range

Found via differential fuzzing

Aditya-9-6:
Rust's `core::fmt` formatting argument precision fits in `u16::MAX` (`65535`). When formatting numbers with precision exceeding 65535 (such as `{{ "%.65536f"|format(1.0) }}`), `format!` panics with `"Formatting argument out of range"`.

This adds bounds checks against `u16::MAX` during format spec parsing and float formatting (including computed decimal places in general format), returning an `InvalidOperation` error instead of panicking.

Fixes #946


mitsuhiko:
Closing as superseded by eaeb3762, which fixes #946 while preserving support for precisions beyond Rust's native formatter limit.

